Issue #26, opened 2023-11-23
[Editor] Model importing scale


michaelsakharov:
FBX files and some others seem to import with drastically weird scales
Seems to be related to: https://github.com/assimp/assimp/issues/2166

10xJosh:
Ill look into this one, hopefully it goes a bit more smoothly than the other bug 😅

michaelsakharov:
Awesome thanks!
It seems Assimp provides some sort of Unit Scale modifier we can use to scale the models on import.

It would be done inside ModelImporter.cs
Probably on the RootNode/Gameobject which is created around 258, just do 
rootNode.Scale = something
Im not sure how exactly to extract the unit scale from assimp though or how it works.

michaelsakharov:
Let me know if you need a some models to test

10xJosh:
> Awesome thanks! It seems Assimp provides some sort of Unit Scale modifier we can use to scale the models on import.
> 
> It would be done inside ModelImporter.cs Probably on the RootNode/Gameobject which is created around 258, just do rootNode.Scale = something Im not sure how exactly to extract the unit scale from assimp though or how it works.
> 
> Reading the issue i linked above it might be something like: scene.Metadata["UnitScaleFactor"].Data;

I tried adding the rootnode stuff but im not sure it worked. I'll have to look more into it tomorrow, I'm a little in over my head here but ill figure it out.

How have you been importing the objects? I've just been putting it in the folder and going into the editor and dragging and dropping. Is there another way to import?

michaelsakharov:
That is the correct way, for now, you might have to restart the editor for it to properly import the model.

michaelsakharov:
Might be more to this then I initially thought.

michaelsakharov:
Added a GlobalScale flag for ModelImporter which may help some cases
Also added a UnitScale modifier value like unity has, this doesn't solve the issue entirely, but it provides a decent enough solution.

PR #61, opened 2023-11-28
Fixed issue Where .fbx files Were Being imported Improperly

10xJosh:
This fix is for issue #26. It turns out the issue is caused by Json deserialization of the object. I wasn't able to pinpoint exactly where at in the deseiralization the scale was becoming changed but the lines of code I implemented should fix the issue. Oddly enough, the 2nd tier child in the hierarchy was where the issue was manifesting from.

In addition to the scale issue, the rotation of the object was off as well. Implemented a fix for that as well

2023-11-27 22_13_28-Prowl

With these changes, it might be better to change up the `EngineObject.Instantiate()` to return a GameObject instead of an engineObject to reduce the amount of casting in the other methods. I noticed the `Instantiate()` method is only being called in the hierarchy so it wouldn't be too hard to implement. Let me know if that's something you want me to implement or if there's anything I'm missing.

### DIFF
--- a/Prowl.Editor/EditorWindows/HierarchyWindow.cs
+++ b/Prowl.Editor/EditorWindows/HierarchyWindow.cs
@@ -100,8 +100,19 @@ public class HierarchyWindow : EditorWindow {
                 if (DragnDrop.ReceiveAsset<GameObject>(out var original))
                 {
                     GameObject clone = (GameObject)EngineObject.Instantiate(original.Res!, true);
+                    clone.Scale = original.Res!.Scale;
                     clone.Position = original.Res!.Position;
                     clone.Orientation = original.Res!.Orientation;
+                    clone.Rotation = original.Res!.Rotation;
+
+                    foreach (var child in clone.Children)
+                    {
+                        child.Scale = original.Res!.Scale;
+                        child.Orientation = original.Res!.Orientation;
+                        child.Position = original.Res!.Position;
+                        child.Rotation = original.Res!.Rotation;
+                    }
+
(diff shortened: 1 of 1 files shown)


Prowl.Editor/EditorWindows/HierarchyWindow.cs
```diff
@@ -100,8 +100,19 @@ protected override void Draw()
                 if (DragnDrop.ReceiveAsset<GameObject>(out var original))
                 {
                     GameObject clone = (GameObject)EngineObject.Instantiate(original.Res!, true);
+                    clone.Scale = original.Res!.Scale;
                     clone.Position = original.Res!.Position;
                     clone.Orientation = original.Res!.Orientation;
+                    clone.Rotation = original.Res!.Rotation;
```
michaelsakharov, line 106: Rotation and Orientation are the same thing just different ways to access them, Orientation is a Quaternion Rotation being a Vector3, so don't need this second assignment.

michaelsakharov: Strange, I looked into the model scale issue a bit this morning and there might be more to it then I thought, and this rotation thing might be a bug, If it's ok with you I might take this over, might be trickier to fully solve all edge cases then I initially thought.

10xJosh: Wow, yeah it does seem like its gonna be a lot more tricky! Go for it, you can have this issue; I'll get started on another issue

PR #90, opened 2024-01-01
simple config dir check

recore67:
Looks bad but, It seems that the script can't find `Project.Projects_Directory` on first build and halts at `File.WriteAllText`. However If you'd manually create the dir, build again, then AGAIN also delete the dir, the saving will work fine forever, so its weird. 

summary:
`File.WriteAllText` can create both the file and directory just not when building first time, for some reason.

### DIFF
--- a/Prowl.Editor/EditorApplication.cs
+++ b/Prowl.Editor/EditorApplication.cs
@@ -49,6 +49,10 @@ public unsafe class EditorApplication : Application {
     {
         string filePath = Path.Combine(Project.Projects_Directory, "EditorConfig.setting");
         string json = JsonSerializer.Serialize(EditorConfig);
+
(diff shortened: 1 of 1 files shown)


Prowl.Editor/EditorApplication.cs
```diff
@@ -49,6 +49,10 @@ public static void SaveConfig()
     {
         string filePath = Path.Combine(Project.Projects_Directory, "EditorConfig.setting");
         string json = JsonSerializer.Serialize(EditorConfig);
+
+        if(!Directory.Exists(Project.Projects_Directory))
```
michaelsakharov, line 53: CreateDirectory itself already does an Exists check before creating so don't need the if check.

michaelsakharov:
Thanks for fixing this, Had no idea this was an issue as I've had the directory for a while now.

PR #101, opened 2024-02-29
File naming proposal

Kuvrot:
### A file naming system proposal
Created a file named "FileNamingWindow.cs" to create a window with text input and buttons and integrate the "create" methods from "MainMenuItem.cs"

gif_3


Prowl.Editor/EditorWindows/FileNamingWindow.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    protected override void Draw()
+    {
+        WindowDrawnThisFrame = true;
+        ImGui.SetWindowSize(new Vector2(480,32));
+        ImGui.InputText("File name", ref createName, 0x100);
+        ImGui.SameLine();
+        if (ImGui.Button("Save file"))
+        {
+            if (fileType == "Script")
```
michaelsakharov, line 45: Instead of hardcoding the types like this, could you pass in an Action to the constructor?
That way this becomes a more generic method that can be reused without changing it.

We could also rename this class to CreateNewFile.cs or something and use it for all file creation

Kuvrot: Sure, I'll work on that.

Kuvrot:
I'm having problems when creating files inside folders, because they are being created at the root folder, not inside the desired folder.

michaelsakharov:
> I'm having problems when creating files inside folders, because they are being created at the root folder, not inside the desired folder.

Is that something being caused by this PR? The code looks correct.

Kuvrot:
No, I think it has to do with the logic but I'm not sure how, if I print the path, it prints the asset folder path not the desired path to create the folder.

michaelsakharov:
I'll merge this for now, let's create an issue for that then.

Issue #103, opened 2024-03-01
[Editor] I'm having problems when creating files inside folders, because they are being created at the root folder, not inside the desired folder.


michaelsakharov:
I'm having problems when creating files inside folders, because they are being created at the root folder, not inside the desired folder.

_Originally posted by @Kuvrot in https://github.com/michaelsakharov/Prowl/issues/101#issuecomment-1972314439_

PR #104, opened 2024-03-02
File creation fixed (Issue  #103)

Kuvrot:
### Fix to the problem where the files were being created only at the root folder instead of the desired one (Issue #103 )

https://github.com/michaelsakharov/Prowl/assets/23508114/8ba3b94d-ce69-4880-802a-e70ce10895cd

I made the CurDirectory variable from the AssetBrowserWindow.cs static in order to indicate the path where the file will be created in MainMenuItems.cs

### DIFF
--- a/Prowl.Editor/EditorWindows/MainMenuItems.cs
+++ b/Prowl.Editor/EditorWindows/MainMenuItems.cs
@@ -16,7 +16,7 @@ namespace Prowl.Editor.EditorWindows
         [MenuItem("Create/Folder")]
         public static void CreateFolder()
         {
-            Directory ??= new DirectoryInfo(Project.ProjectAssetDirectory);
+            Directory ??= AssetBrowserWindow.CurDirectory;
             var folderAction = new Action<string, string>((directory, createName) =>
             {
                 DirectoryInfo dir = new DirectoryInfo(Path.Combine(directory, createName));
@@ -32,7 +32,7 @@ namespace Prowl.Editor.EditorWindows
         [MenuItem("Create/Material")]
         public static void CreateMaterial()
         {
-            Directory ??= new DirectoryInfo(Project.ProjectAssetDirectory);
+            Directory ??= AssetBrowserWindow.CurDirectory;
             var materialAction = new Action<string, string>((directory, createName) =>
             {
                 Material mat = new Material(Shader.Find("Defaults/Standard.shader"));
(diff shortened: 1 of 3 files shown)


Prowl.Editor/EditorWindows/MainMenuItems.cs
```diff
@@ -16,7 +16,7 @@ public static class MainMenuItems
         [MenuItem("Create/Folder")]
         public static void CreateFolder()
         {
-            Directory ??= new DirectoryInfo(Project.ProjectAssetDirectory);
```
michaelsakharov, line 19: Hmm, these null checks here were placed as a fallback when no AssetBrowserWindow has been created.
AssetBrowserWindow.CurDirectory can be null
It also should support when right-clicking a folder and pressing Create Material, it should create it inside the folder not the directory I'm inside.

michaelsakharov: The problem is further up, Whatever is calling MenuItem.DrawMenuRoot("Create");
Isnt setting the MainMenuItems.Directory properly, which at a first glance looks like thats the case in a few issues.

michaelsakharov:
Pushed a possible fix to this just now, Let me know if it works.

Prowl.Editor/EditorWindows/MainMenuItems.cs
Kuvrot: okok, I get it now.

Kuvrot:
I didn't work for me.

michaelsakharov:
Dang, Can you show a clip of how exactly your creating the files?

michaelsakharov:
Sorry i missed a compile error

Kuvrot:
> It also should support when right-clicking a folder and pressing Create Material, it should create it inside the folder not the directory I'm inside.

To be more specific, this is working, right clicking a folder and creating a file, will create the file inside the folder clicked.

What is not working is this: 

https://github.com/michaelsakharov/Prowl/assets/23508114/bc0b4880-354c-418c-ad73-236094a462d8

michaelsakharov:
> > It also should support when right-clicking a folder and pressing Create Material, it should create it inside the folder not the directory I'm inside.
> 
> To be more specific, this is working, right clicking a folder and creating a file, will create the file inside the folder clicked.
> 
> What is not working is this:
> 
>  Prowl2024-03-0217-41-29-ezgif.com-crop-video.mp4

This seems to be working for me on my last commit

Kuvrot:
Oh really?, let me check it again then.

Kuvrot:
Ahhhh yes, sorry you are right, I made a new clone of the repo and it's working.

michaelsakharov:
Awesome :D

PR #106, opened 2024-03-10
Added selected GameObject focus upon pressing "F" key in the viewport window

jihadkhawaja:


Prowl.Editor/EditorWindows/ViewportWindow.cs
```diff
@@ -390,6 +390,14 @@ protected override void Update()
                 else if (Input.GetKeyDown(Key.T)) GizmosOperation = ImGuizmoOperation.Bounds;
 
             }
+
+            if (Input.GetKeyDown(Key.F) && HierarchyWindow.SelectHandler.Selected.Any())
+            {
+                var selectedWeakGO = HierarchyWindow.SelectHandler.Selected.First();
```
michaelsakharov, line 396: Rather than just the first
Can we do something like:
`
// Calculate the average position of all selected GameObjects
Vector3 averagePosition = Vector3.zero;
foreach (var obj in HierarchyWindow.SelectHandler.Selected)
    if (obj.Target is GameObject go)
        averagePosition += go.transform.position;
averagePosition /= HierarchyWindow.SelectHandler.Selected.Count;
`
        
This way it supports multiple objects selected. Might wanna have some dynamic distance rather than just 3f, but just centering works fine for now.

PR #141, opened 2024-08-02
Added icon to editor and improved Tooltip display

treviasxk:
### Tooltips
• Now it will adjust within the editor window if the message is too large and exceeds the limits.
• Fixed background tooltip size.
• Created enum TooltipAlign, and added param Align in Tooltip function to align display.
• Add tooltips in buttons console.
• Remove border for a more minimalist design

### Editor
• Added prowl icon to file executable and editor window.
_It was necessary to add the SixLabors.ImageSharp package to the project in order to load the image into Silk.Net. [Code reference](https://github.com/dotnet/Silk.NET/blob/b079b28cd51ce447183cfedde0a85412b9b226ee/src/Lab/Experiments/BlankWindow/Program.cs#L82)._

### Script
• Updated UpdateSubmodules.bat, old command was not downloading the submodules.

https://github.com/user-attachments/assets/b903e947-ee82-426f-a67f-eb68a91eb7bb


michaelsakharov:
Thanks! I like the new design for the Tooltips and the clamping of the tooltip inside the window is a nice addition!

The icon for the executable is nice, however, adding ImageSharp to the runtime isn't ideal as it adds significant size to the runtime.
So we would need to find another solution for that.
Also gui.Tooltip(); is designed to be called after an interactable node.
So it should be moved down a little in StyledButton to be after the gui.Node().Enter().

Could you also rebase this to the Veldrid branch? The main branch is a bit behind at the moment.

treviasxk:
I just added it to the veldrid #142 branch, you can close this pull here to avoid conflicts.

In the new pull I kept the gui.Tooltip(); I kept the original positions in the code execution.

PR #142, opened 2024-08-08
Updated Tooltip display and added icon to editor - Veldrid

treviasxk:
### Tooltip
• Now tooltip clamp position inside editor.
• Remove border for a more minimalist design.
• Improved tooltip background position.
• Added enum TooltipAlign in method Tooltip() to align in display.
• Added Tooltip in console buttons and play/pause/stop editor button.
• The calculation was improved in relation to pull #141 of the main branch.

### Editor
• Added icon to the editor executable.
_This time I didn't need to use the SixLabors.ImageSharp package, Veldrid already included the executable icon in the window. (no more resource consumption)_

Logo

### Demonstration video
In the demo, I placed text with different alignment on each console button as an example of displaying and showing the behavior of the tooltip on each edge of the editor window.

https://github.com/user-attachments/assets/26bf9b3b-5e51-4d61-86f3-9182513159fe

Note: the Veldri branch is not rendering the graphics correctly on my computer. OpenGL displays a black screen with some red textures, Vulkan and DirectX11 display the GUI but crash if I select the viewport or game view window.

But I was still able to place the tooltips with DirectX11


michaelsakharov:
Awesome thanks!
The rendering is a bit broken in this branch at the moment as it's still under development.
OpenGL should work though as that's the one I use the most weird...

Issue #154, opened 2024-09-13
Code style: file scoped namespace


brmassa:
Newer versions of C# allow file-scoped namespaces. Instead of creating basically double brackets to accommodate namespace + class.

```cs
using Sytem;

namespace Prowl 
{
    public class MyClass
    {
        public void MyMethod()
        {
            DoStuff();
        }
    }
}
```

become

```cs
using Sytem;

namespace Prowl;

public class MyClass
{
    public void MyMethod()
    {
        DoStuff();
    }
}
```

## Advantages:

- Less indentation
  - easier to read on Discord or narrow chats
  - most IDEs try to split lines above 80 character
- Cleaner code
- Most of the conversion is done only once

## Disadvantages:

- HUGE diff when updating all files, since most of the code base was using the old style. The commit will alter basically all lines of all files.

---

Because one might be annoyed, it's recommended to do during a big rewrite of the code, like the current `development` branch. I suggest to each developer to express opinion.

# # Tasks

quite simple

- [x] Convert all files to file scoped namespaces
- [x] include a rule inside `.editorconfig` to raise warnings when using normal brackets

michaelsakharov:
Im down for this, The reduction in Indentation would be nice.
Especially for UI stuff, since the UI API generally results in very long lines.

brmassa:
done

PR #155, opened 2024-09-14
Chore: file scoped namespace

brmassa:
Implementation of #154 

Note: External projects (submodules DotCast and Veldrid) were not affected. It will raise warnings until they are changed in their own repos


michaelsakharov:
Is merging into Main intended?

brmassa:
Well why not?

AFAIK, development branch will basically overwrite main.

If eventually there is any file from main that will not be overwritten, we can perform this change on these dozen remaining files later.

Any thoughts?

PaperPrototype:
I think the plan was to do everything in development and once it is all cleaned up and ready just do a single merge into main

brmassa:
@michaelsakharov ohhh now that Ive noticed that generate the Merge Request against main. My bad. It's designed to be applied to dev.

PR #157, opened 2024-09-16
Fix: never compare equality of floating numbers

brmassa:
`single` and `double` floats use an especial arithmetic that uses a trade-off of being super big/small, but not very precise.

Eventually, 100f - 10f might result in `90.000000001` instead of pure `90f`, so when comparing if 2 floats are the same, we need actually to check if they are close enough. Doing the pure comparison might result in weird behaviors.

- [ ] decide a max difference that will still be considered the same numbers (the code uses `0.0001f`)
- [ ] create a threshold number (Application.FloatEqualThreshold ?)
- [ ] convert all float number tests to use this constant (generally, get the absolute difference and check if is smaller)

### DIFF
--- a/Prowl.Runtime/Components/Physics/Types/BepuNarrowPhaseCallbacks.cs
+++ b/Prowl.Runtime/Components/Physics/Types/BepuNarrowPhaseCallbacks.cs
@@ -49,7 +49,7 @@ public unsafe struct BepuNarrowPhaseCallbacks : INarrowPhaseCallbacks
         var b = CollidableMaterials[pair.B];
         pairMaterial.FrictionCoefficient = a.FrictionCoefficient * b.FrictionCoefficient;
         pairMaterial.MaximumRecoveryVelocity = (float)MathD.Max(a.MaximumRecoveryVelocity, b.MaximumRecoveryVelocity);
-        pairMaterial.SpringSettings = Math.Abs(pairMaterial.MaximumRecoveryVelocity - a.MaximumRecoveryVelocity) < Application.FloatEqualThreshold ? a.SpringSettings : b.SpringSettings;
+        pairMaterial.SpringSettings = MathD.ApproximatelyEquals(pairMaterial.MaximumRecoveryVelocity, a.MaximumRecoveryVelocity) ? a.SpringSettings : b.SpringSettings;
         //For the purposes of the demo, contact constraints are always generated.
         ContactEvents.HandleManifold(workerIndex, pair, ref manifold);
         Physics.Characters.TryReportContacts(pair, ref manifold, workerIndex, ref pairMaterial);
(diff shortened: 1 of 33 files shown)


Prowl.Runtime/Components/Physics/Types/BepuNarrowPhaseCallbacks.cs
```diff
@@ -48,7 +49,7 @@ public unsafe bool ConfigureContactManifold<TManifold>(int workerIndex, Collidab
         var b = CollidableMaterials[pair.B];
         pairMaterial.FrictionCoefficient = a.FrictionCoefficient * b.FrictionCoefficient;
         pairMaterial.MaximumRecoveryVelocity = (float)MathD.Max(a.MaximumRecoveryVelocity, b.MaximumRecoveryVelocity);
-        pairMaterial.SpringSettings = pairMaterial.MaximumRecoveryVelocity == a.MaximumRecoveryVelocity ? a.SpringSettings : b.SpringSettings;
+        pairMaterial.SpringSettings = Math.Abs(pairMaterial.MaximumRecoveryVelocity - a.MaximumRecoveryVelocity) < Application.FloatEqualThreshold ? a.SpringSettings : b.SpringSettings;
```
michaelsakharov, line 52: I imagine in the case of BepuPhysics this wouldn't be ideal due to possible performance costs it would come with.
This is very frequently called, and it wouldnt matter much if there are errors due to floating precision here.

michaelsakharov: However, now that I look more closely at the rest of the method, there's a hefty float/double cast and MathD.Max being used. So maybe it doesn't matter as much as I thought.

PaperPrototype:
instead of a threshold maybe create a custom comparison check? That uses `Application.FloatEqualThreshold`
- [ ] Math.FloatIsEqual(floatValue, comparedTo)
- [ ] Math.FloatIsGreater(floatValue, comparedTo)
- [ ] Math.FloatIsLesser(floatValue, comparedTo)

michaelsakharov:
> instead of a threshold maybe create a custom comparison check? That uses `Application.FloatEqualThreshold`
> 
> * [ ]  Math.FloatIsEqual(floatValue, comparedTo)
> * [ ]  Math.FloatIsGreater(floatValue, comparedTo)
> * [ ]  Math.FloatIsLesser(floatValue, comparedTo)

This is a good point Unity has Mathf.ApproximatelyEquals, Since our goal is to sorta mimic Unity's API to make transitioning as seamless as possible, it makes sense to recreate that idea.

brmassa:
I agree with all comments:

* Bepu: the new check does not add much calculation. Also, pure equality might create weird physic bugs, very had to track
* `Mathf.ApproximatelyEquals`: good option. Also, I've arbitrarily chosen 0.0001f. It would be nice to check the value selected by Unity
* `Math.*` methods make it more readable. a* what about using `Mathf.*`? b* A operator `==` override (for floats and doubles)?

brmassa:
I've created Mathf.ApproximatelyEquals because it's the same method name used for MathD.

Also, I replaced several Epsilon comparisons with the Mathf.Epsilon and MathD.Epsilon

Issue #158, opened 2024-09-16
Fix loads of build warnings #development


brmassa:
When building `development`, the compiler raises loads of warnings (500+). There are problems are well known sources of bugs, like not checking for `null`.

This issue will centralize all initiatives of fixing the warnings. Warning codes sorted by frequency.

- [ ] CS8602: 158
- [ ] CS8618: 102
- [ ] CS8604: 61
- [ ] CS8600: 59
- [x] CS8625: 16
- [ ] CS1030: 11
- [x] CS0618: 11
- [x] CS8605: 11
- [ ] CS8601: 10
- [x] CS8603: 10
- [x] CS8629: 9
- [ ] CS0169: 9
- [x] CS0219: 5
- [x] CS0162: 5
- [x] CS0108: 4
- [x] CS4014: 4
- [x] CS9191: 3
- [x] CS8619: 3
- [x] CS0649: 3
- [x] CS8765: 2
- [x] CS8767: 2
- [x] CS8620: 2
- [ ] CS0067: 2
- [x] CS0659: 1
- [x] CS0661: 1
- [x] CS8714: 1
- [x] CS8607: 1
- [ ] SYSLIB0014: 1
- [x] CS8509: 1
- [x] CS1998: 1

Also, for the sake of a better code quality, we might add more rules on `.editorconfig` to enforce good practices.

PR #159, opened 2024-09-16
Fix/build warnings 2024 09 15

brmassa:
Attacking #158

fixes for:

CS0108: 4
CS4014: 4
CS9191: 3
CS8619: 3
CS0649: 3
CS8765: 2
CS8767: 2
CS8620: 2
CS0659: 1
CS0661: 1
CS8714: 1
CS8607: 1
CS8509: 1
CS1998: 1

from 529 to 467 compilation warnings (not considering `DotCast` nor `Veldrid`)

> NOTE: If one wants to focus on solving these issues, I suggest to adding
 `dotnet_diagnostic.<CODE>.severity = error`
 inside 
`.editorconfig`. 
Your IDE might immediately point to errors. It not, try compiling the code and check the log

### DIFF
--- a/Prowl.Editor/Assets/AssetDatabase.Core.cs
+++ b/Prowl.Editor/Assets/AssetDatabase.Core.cs
@@ -140,7 +140,7 @@ public static partial class AssetDatabase
                         if (ProcessFile(file, out _))
                             toReimport.Add(file);
                     }
-                    else if (!assetPathToMeta.TryGetValue(file, out var meta))
+                    else if (!assetPathToMeta.TryGetValue(file, out _))
                     {
                         // File hasent changed but we dont have it in the cache, process it but dont reimport
                         Debug.Log("Asset Found: " + file);
@@ -169,7 +169,7 @@ public static partial class AssetDatabase
                 cacheModified = true;
                 bool hasMeta = assetPathToMeta.TryGetValue(file, out var meta);
 
-                if (hasMeta)
+                if (hasMeta && meta is not null)
                 {
                     assetPathToMeta.Remove(file);
 
@@ -352,11 +352,12 @@ public static partial class AssetDatabase
             Debug.LogError($"No valid meta file found for asset: {ToRelativePath(assetFile)}");
             return false;
         }
-        if (meta.Importer == null)
-        {
-            Debug.LogError($"No valid importer found for asset: {ToRelativePath(assetFile)}");
-            return false;
-        }
+        // TODO: FIXME: its always not null
+        // if (meta.Importer == null)
+        // {
+            // Debug.LogError($"No valid importer found for asset: {ToRelativePath(assetFile)}");
+            // return false;
+        // }
 
         // Import the asset
         SerializedAsset ctx = new(meta.guid);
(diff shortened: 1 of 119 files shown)


brmassa:
CS8629 and some other fixes

brmassa:
CS8603 and some CS8601

down to 406 warnings

brmassa:
down to 361 warnings

michaelsakharov:
It might make more sense to disable the Readonly warning.
As there's a ton of cases with the Readonly's where it doesn't make a whole lot of sense.
Including a lot of public API intended for end users to change in runtime

michaelsakharov:
The issue with switching to Properties is that moves away from the Unity API. Unity doesn't really even support properties-only fields, our serializer also doesn't handle them properly and it would be a reasonably big change for newcomers to c# who only know c# through Unity.

I think the read-only one makes sense to remove for the time being, What do you think @sinnwrig

PaperPrototype:
hehe yeah im one of those unity C# only people :P

PaperPrototype:
One thought I have is maybe not worrying about Unity only devs? We would only be perpetuating the problem.

brmassa:
It's problem going to bite us in the future, but for the time being, I'm reverting it. There are still a lot (200+) other warning to attack.

michaelsakharov:
> One thought I have is maybe not worrying about Unity only devs? We would only be perpetuating the problem.

I mean a huge part of prowl is to be a unity alternative, so it's main audience by far will be unity devs. On top of the fact that it's not really a problem to perpetuate in the first place. Its less a problem and more a design choice.

brmassa:
ok. Commit reverted.

I believe Unity did some choices back in old days due lack of resources. Joachim repeatedly said there are several design mistakes that now they kinda have to live with.

Their serializer is sub-par. Sirenix (Odin) serializer is much more capable.

Using public mutable fields is a very bad practice. Basically all Static Analyzers will complain: Visual Studio, Rider, Codacy, Qodana, etc.

Sicne drag-and-drop replacement is virtually impossible, I believe we could have peace of mind on taking some things in other direction. As long it's well documented and explained to users.

sinnwrig:
> The issue with switching to Properties is that moves away from the Unity API. Unity doesn't really even support properties-only fields, our serializer also doesn't handle them properly and it would be a reasonably big change for newcomers to c# who only know c# through Unity.
> 
> I think the read-only one makes sense to remove for the time being, What do you think @sinnwrig

The readonly warnings should only be appearing because of C# style recommendations, correct? If so, it would probably be feasible to revert the warnings back to suggestions in the .editorconfig, since the compiler can't be aware of how a user and the serializer might interact with it. On top of that, the warning seems a bit like premature optimization, since the performance boost gained by setting fields to readonly is negligible in most cases unless it's in perf. critical code, in which case it can be optimized manually later down the line.

michaelsakharov (changes requested):

sinnwrig:
> ok. Commit reverted.
> 
> I believe Unity did some choices back in old days due lack of resources. Joachim (Unity ttech chief back then) repeatedly said there are several design mistakes that now they kinda have to live with.
> 
> Their serializer is sub-par. Sirenix (Odin) serializer is much more capable.
> 
> Using public mutable fields is a very bad practice. Basically all Static Analyzers will complain: Visual Studio, Rider, Codacy, Qodana, etc.
> 
> Sicne drag-and-drop replacement is virtually impossible, I believe we could have peace of mind on taking some things in other direction. As long it's well documented and explained to users.

On the public mutable fields, it seems a bit harsh to remove them completely because of bad design practice, since for many cases, having simple, raw, exposed values without a get/set (even if it's auto-implemented) is useful. Worst-case, the [field:SerializeField] attribute can be used to let the serializer get to an auto-implemented property.

Issue #151, opened 2024-09-13
[Editor] CLI commands


brmassa:
It would be nice to have a series of options and arguments when calling Prowl.

Features:

- [ ] Open project: `prowl --project PATH`
- [ ] Create project: `prowl new --project PATH`
- [ ] Build project: `prowl build --project PATH --output PATH`
- [ ] Version: `prowl version`

references:

- https://docs.unity3d.com/Manual/CommandLineArguments.html
- https://docs.flaxengine.com/manual/editor/advanced/command-line-access.html

brmassa:
in #152 , ive used the term `create` instead `new`.

NOTE: consider adding a local README.md inside `Prowl.Editor/Editor/CLI` explaining all the command line commands for reference.

brmassa:
`prowl version` (`prowl.exe version`) implemented automatically by `CommandLineParser`

PaperPrototype:
@brmassa The latest branch as of now is "New-Editor-UI" I think something like this should be significantly easier now with the latest architecture.

sinnwrig:
All build options (and more) have been implemented in New-Editor-UI branch.

```
--help 
--project
--restore-scene
--buildmode
--output
```

PaperPrototype:
Beautiful

PR #175, opened 2024-09-17
Feat: CLI `build`  command (WIP)

brmassa:
#151

Build your app using the command line `prowl build --project <PATH>`. It will invoke the build pipeline.

Relevant changes:

- [ ] Program.cs: BuildCommand to prepare and invoke the build pipeline
- [ ] ConsoleWindown and Debug: allowing colored messages on build
- [ ] ScriptableSingleton.cs: allow to parse FilePathAttribute.Location.EditorSetting when not using the Editor
- [ ] Miscelaneous code style changes

### DIFF
--- a/Prowl.Editor/Build/Desktop_Player.cs
+++ b/Prowl.Editor/Build/Desktop_Player.cs
@@ -7,153 +7,152 @@ using Prowl.Editor.Assets;
 using Prowl.Runtime;
 using Prowl.Runtime.Utils;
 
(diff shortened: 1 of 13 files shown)


brmassa:
Currently, is failing to load the Scenes from `BuildProjectSetting.Instance.Scenes`. It tries to call the `AssetRef`, them try to ready it's `.Res`, which does not have GUID so it instanciate an empty Scene.

the most important method of this MR is `Program.BuildCommand`

PR #179, opened 2024-09-18
Feat: Prowlhash test suite

brmassa:
Prowlhash

- Code optimization to avoid repetition (prone to errors)
- Prowlhash Seed can be manually set
- replace GetHashCode for a stable hash generator because GetHashCode is unpredictable on each run, Os and archteture.
- 20+ unit tests to ensure it's quality

dependent on #177

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 # This workflow will build a .NET project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net
 
-name: Build
+name: Build and Test
 
 on:
   push:
@@ -9,7 +9,6 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
@@ -31,3 +30,7 @@ jobs:
       # Build
       - name: Build
         run: dotnet build --no-restore
+
+      # Test
+      - name: Test
+        run: dotnet test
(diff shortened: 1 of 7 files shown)


michaelsakharov:
Could you explain the StableHash a bit more?
Does this not expect values to always be strings, and how does it handle class references?
GetHashCode is nice since two classes that are different instances return different HashCodes unless they override the function.

brmassa:
**GetHashCode**: The default GetHashCode method in .NET can produce different hash codes for the same object in different runs of the application. This is because the default implementation may use the object's _memory address_, which can/will change between runs.

**StableHash**: The StableHash method aims to produce the same hash code for the same object across different runs. This is achieved by using a deterministic seed and by hashing the object's state rather than its memory address

michaelsakharov:
If im not mistaken StableHash works by the objects ability to convert to string?
What about like classes that don't implement ToString()? It would fallback to the Type name and all instances would receive the same hash? And what if the implementation of ToString() isn't reliable for object's state?

Im not entirely against this, just curious how this method behaves compared to the default GetHashCode.
If that is the case it might be nice to support both, like GetHashCode(bool deterministic) or something, as I can see a few cases where just GetHashCode might be nice to have for having a unique "hash" per instance of a class (the memory address).

michaelsakharov:
Also if you don't mind me asking whats your original reasoning/intention for having a deterministic hashcode?

PR #191, opened 2024-11-12
Vector3.cs - Calculate Perpendicular vector

PaperPrototype:


michaelsakharov:
I think having a built in Create Perpendicular isn't too needed, As that's what a cross is already made todo.

PaperPrototype:
Any chance we could keep it? I know what a Cross is for but I can't tell you how many times I needed an arbitrary Perpendicular vector from a single vector input.

PR #197, opened 2024-12-10
Build fixes

ZeppelinGames:
- Fixed BuildWindow crash due to substring name missing char
- Added `GetDefaultAssets` function to `AssetDatabaseUtils`
- Added default assets to be packed in Desktop compilation. Fixes building projects that use default assets and had asset packing set to 'used assets'
- Uniformity on project window loading overlay (doubleclicking project now does same action as clicking 'open' button)

### DIFF
--- a/Prowl.Editor/Editor/ProjectsWindow.cs
+++ b/Prowl.Editor/Editor/ProjectsWindow.cs
@@ -252,7 +252,7 @@ public class ProjectsWindow : EditorWindow
         // Text pos + offset/padding
         Vector2 openInfoTextPos = footer.Position + new Vector2(8f, 8f);
         gui.Draw2D.DrawText($"Opening '{SelectedProject.Name}'...", openInfoTextPos);
-        
+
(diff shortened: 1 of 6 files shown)


PaperPrototype:
yay

Prowl.Editor/Editor/ProjectsWindow.cs
```diff
@@ -469,8 +497,21 @@ private void DrawProjectContextMenu(Project project)
         {
             using (popupHolder.Width(180).Padding(5).Layout(LayoutType.Column).Spacing(5).FitContentHeight().Enter())
             {
+                // Getting Interactable (IsHovered) removes interaction from styled buttons?
```
michaelsakharov, line 500: This is a bit odd, IsHovered and Interactables are just rectangles, this should work just fine with IsHovered?
Could you elaborate on what happened here? Did interaction just completely fail?

Other then this (which isn't really a problem preventing merging) everything looks good!

PR #227, opened 2025-10-13
Fixed gltf 2.0 bug

EJTP:


michaelsakharov:
that check seems a bit fishy, I think it would make more sense to just make an Issue on this problem, and because its not a problem on our end, so fixing it on our end is a hacky band aid.

We need to make and maintain our own Assimp.Net fork or port Assimp to c#

EJTP:
True

PR #235, opened 2026-01-03
Change DLLImport to newer LibraryImport for miniaudio

Tryibion:
This uses the newer library import instead of the dllimport for the miniaudio library.

I did not convert all of the callbacks because those are handled differently using LibraryImport where they use a delegate pointer, but this is a decent start and the audio demo still works.


michaelsakharov:
Woops, didnt mean to press that.

Are you looking to have this reviewed/merged now? Or is this still a draft?

Tryibion:
It can be reviewed and merged. The only caveat is if you notice some of the callbacks are not working they will just need to be changed as I only changed the initialize one since that was the only one the audio demo seemed to be using.

PR #244, opened 2026-04-11
Font switching support and Project settings panel tweaks

xZekro51:
- Actually added Font switching support in themes (missed in [PR#243](https://github.com/ProwlEngine/Prowl/pull/243))
- Tweaked project settings panel with padding and navbar spacing

### DIFF
--- a/Prowl.Editor/Panels/ProjectSettingsPanel.cs
+++ b/Prowl.Editor/Panels/ProjectSettingsPanel.cs
@@ -15,6 +15,7 @@ public class ProjectSettingsPanel : DockPanel
 
     private int _selectedIndex;
 
+
     public override void OnGUI(Paper paper, float width, float height)
     {
         var font = EditorTheme.DefaultFont;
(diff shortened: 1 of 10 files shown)


Prowl.Editor/Panels/ProjectSettingsPanel.cs
```diff
@@ -57,10 +59,10 @@ public override void OnGUI(Paper paper, float width, float height)
                     string icon = string.IsNullOrEmpty(entry.Icon) ? EditorIcons.Gear : entry.Icon;
 
                     paper.Box($"ps_cat_{i}")
-                        .Height(26).ChildLeft(8).Rounded(3)
+                        .Height(30).ChildLeft(8).Rounded(3).Margin(0,0,0,new UnitValue(Units.Pixels, EditorTheme.VerticalNavbarSpacing))
```
michaelsakharov: You should be able to just pass in EditorTheme.VerticalNavbarSpacing? Its automatically casted to a Unitvalue of pixels, so you don't need to use the constructor.

michaelsakharov: other then that this looks great! Thanks so much for the help, UI cleanup like this is exactly what I need the most right now!

xZekro51: You're right! I'm still getting the hang of how Paper works, so things like this are easily missed- I'll update that part right now!

michaelsakharov: All good, If you run into anything your unsure about in paper, don't be afraid to spam me with questions on Discord! Only thing at the moment worth noting is to avoid using ChildLeft, Right, Top, Bottom, i plan to deprecate those and build proper Padding support their a bit strange at the moment. its just low priority right now, trying to pump out the rest of the editor.

PR #250, opened 2026-04-15
Added Editor Window State saving

xZekro51:
Implemented the Editor Settings saving the window position, size, and maximization state. It automatically restores that on startup and is shared across projects.


PaperPrototype:
:eyes:

michaelsakharov:
Do we not already have editor layout saving or am I misunderstanding this?

xZekro51:
This solves a different issue- Previously, when the editor was launched, it always defaulted to a resolution (I think it was 1200x800) in the same position of the screen, in a floating window. Now the editor stores the last saved window position, size, and maximization state so that it's restored at startup. It's different from the normal layout saving, which saves the layout of the panels inside the editor.

PR #256, opened 2026-04-19
Reworked RenameOverlay for better renaming support

xZekro51:
Added autofocus upon Rename Overlay opening and positional override. Currently waiting for better (or open) cursor support to automatically select the entire text when starting the rename process (currently, everything on Paper's side to manage that is private- is it fine if I change that as well?)


michaelsakharov:
i think i did add the SelectAllText in paper, i'll check that

PR #258, opened 2026-04-20
Update DesktopBuildPipeline.cs

PaperPrototype:


michaelsakharov:
Need to find a way to automatic this i think... its gonna be annoying.

PR #263, opened 2026-04-25
Build flow rework, Build Project Panel and per-platform build profiles

xZekro51:
Reworked the build process to be more consistent and predictable, as well as strengthening the built project's Program.cs with Assembly probing in the runtimes folder.

Also added provisional PublishTrimmed optional parameter, which slightly reduces build size (I've seen a 50MB reduction, but other settings could be more impactful, even though potentially unsafe).

Made BuildSettings serialized through `Prowl.Echo` to fully support polymorphic lists, as `System.Text.Json`'s way of doing it is way too hacky.

Added `PlatformBuildProfile` with Platform-specific settings for builds, as well as scripting define symbols. `BuildSettings` still holds the general settings that are relevant across different platforms.

Lastly, added support for cmd builds and async builds with progress reporting, which are especially useful for CI/CD systems where builds need to be updated.

Any feedback is welcome!


xZekro51:
Added a screenshot of the new Build Project Panel
image

PR #262, opened 2026-04-24
Collider debug gizmos

PaperPrototype:
- [x] Prowl.Runtime/Components/Physics/Colliders/BoxCollider.cs
- [x] Prowl.Runtime/Components/Physics/Colliders/SphereCollider.cs (not tested)
- [x] Prowl.Runtime/Components/Physics/Colliders/ConeCollider.cs (not tested)
- [x] Prowl.Runtime/Components/Physics/Colliders/CylinderCollider.cs (not tested)
- [x] Prowl.Runtime/Components/Physics/Colliders/CapsuleCollider.cs (not tested)
- [x] Concave MeshCollider (Prowl.Runtime/Components/Physics/Colliders/MeshCollider.cs)
- [x] Convex MeshCollider (Prowl.Runtime/Components/Physics/Colliders/MeshCollider.cs)

This solves https://github.com/ProwlEngine/Prowl/issues/259

Convex mesh colliders work, but I can't seem to get the debug gizmo for convex version working:

Screenshot 2026-04-24 at 5 32 43 AM

---

If you need a project to test with you can download this and open it in prowl: https://github.com/PaperPrototype/KayKitDemos/blob/main/Assets/Paper/Platformer/README.md

PR #266, opened 2026-04-27
IDebugDrawer gizmos for all colliders

PaperPrototype:
Adds debug drawer drawing for Colliders. This was a one line change since support for it was already added previously ~~but for some reason it causes 2 fps~~. Though... FPS drops to 45 fps from 300 fps in my test project.

`Prowl.Runtime/Components/Physics/Colliders/Collider.cs`:

```cs
public override void DrawGizmos()
{
    if (_attachedBody == null || _attachedBody.Handle.IsZero) return;

    _attachedBody.DebugDraw(JitterGizmosDrawer.Instance);
}
```

> [!NOTE]
> You do have to click play for the gizmos to start drawing.

### DIFF
--- a/Prowl.Runtime/Components/Physics/Colliders/Collider.cs
+++ b/Prowl.Runtime/Components/Physics/Colliders/Collider.cs
@@ -340,6 +340,13 @@ public abstract class Collider : MonoBehaviour
         }
     }
 
+    public override void DrawGizmos()
+    {
+        if (_attachedBody == null || _attachedBody.Handle.IsZero) return;
+
+        _attachedBody.DebugDraw(JitterGizmosDrawer.Instance);
+    }
+
     public override void OnValidate()
     {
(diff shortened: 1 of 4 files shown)


PaperPrototype:
This is what it looks like. Though you do have to click play for it to show up in the scene view
Screenshot 2026-04-27 at 3 19 30 AM

michaelsakharov:
Huh we were gonna just remove this you made a better one that works in editor?

PaperPrototype:
Uh no, it only works if you click play and then switch to scene view. I think we should just remove IDebugDrawer. Only problem is I don't know how much better it is when it comes to joints and stuff. It might have utility in those areas since we would have to write custom gizmos for each joint type ourselves manually. Regardless this performance is just abysmal...

PaperPrototype:
Closed in favor of #262

PR #267, opened 2026-04-27
DPI fix (part 2)

PaperPrototype:
Still works on windows just need to test on macos


PaperPrototype:
Only does "TryGetContentScaleViaProc" for windows

PaperPrototype:
Okay confirmed it still works on macos

PR #269, opened 2026-04-30
Mesh collider physics crash

PaperPrototype:
100% vibe coded with claude but it fixes mesh colliders crashing the engine in 2 cases:

- Adding a mesh collider via a script creating a gameobject at runtime
- Mesh collider component setup in a scene


PaperPrototype:
Also for some reason the performance is quite bad, it takes a solid minute to load a mesh collider for my terrain 💀

PR #270, opened 2026-05-01
Rename ReferenceOpenerService to EditorUtils

PaperPrototype:


PaperPrototype:
Or maybe EditorService.cs would be better? @michaelsakharov

PR #273, opened 2026-05-03
Add manual mesh baking option to mesh collider

PaperPrototype:
The mesh colliders currently iterate all triangles in your mesh and create a JitterPhysics TriangleMesh. This can be quite expensive to do for very large meshes. This fixes that and lets you called `BakeMesh` and `SetTriangleMesh` yourself in a thread or C# Task. I'm currently using this PR for my Voxel Terrain project and it removed the stutter I was experiencing before.

### DIFF
--- a/Prowl.Runtime/Components/Physics/Colliders/MeshCollider.cs
+++ b/Prowl.Runtime/Components/Physics/Colliders/MeshCollider.cs
@@ -21,6 +21,7 @@ public sealed class MeshCollider : Collider
 {
     [SerializeField] private AssetRef<Mesh> mesh;
     [SerializeField] private bool convex = false;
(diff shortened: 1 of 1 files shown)


PaperPrototype:
@michaelsakharov when you get the chance can you review this PR?

michaelsakharov:
this should be moved maybe to Mesh Features or something, doing it like this is slow because every Instance of a Mesh Collider of the same mesh has to generate its own TriangleShape So it should store per Mesh, and maybe have like a Physics.BakeMesh(myMesh) so you can generate it for a mesh yourself before you assign it to a mesh collider and do it on another thread if you want.

michaelsakharov:
Done in the New Editor UI Branch

PR #274, opened 2026-05-06
Reworked File creation flow

xZekro51:
Reworked File creation flow to wait for a filename before creating the file. Furthermore, Shaders and Scripts templates were moved from code to template files that could support syntax highlighting, as well as API in EditorApplication to easily retrieve these files.

If anything is wrong, please share feedback on this PR to improve it :pray:


PaperPrototype:
Ah yes this was always bothering me, very nice QOL

PR #276, opened 2026-06-06
[Feature] [UNFINISHED] Added GameObject-based UI

xZekro51:
Added GameObject-based UI to the engine, with support for three types of canvas rendering: Screen Space (on a camera's near plane), Overlay (blit to screen after all rendering is done), World Canvas (treated as normal meshes)

Note that this is still not finished/polished and is lacking lots of default components such as sliders or layout groups.


michaelsakharov:
This work has been merged into the New Editor UI branch.

PR #278, opened 2026-06-16
Moved build to separate thread, improvement to ALC by better sweeping and loose-ends clearing

xZekro51:
Moved the build process to a separate thread when AssetAsyncLoading is enabled in the project settings, improved the ALC Hotload by adding various helpers to sweep the Editor state, removing any loose ends, and adding Selection restoration via token.

Note: ALC reload still won't work in case the Editor is run in Debug mode from VS/Rider, as they keep the ALC alive through the session.


michaelsakharov:
Regarding all the ClearCache stuff, could we make like a Attribute like OnAssemblyUnload that you can add to a static method that automatically registers it and gets cleared on assembly unload, we could add one for assembly unload/load and also a OnScriptCompile for a more generalized one when scripts start compiling, saves a lot of ClearCache methods all over the place needing to be called from one area.

xZekro51:
Agree with your previous comment completely, I just pushed the changes- Added three attributes (OnAssemblyLoad, OnAssemblyUnload, and OnScriptCompile) that are to be assigned to static parameterless methods, which would then be called at the appropriate points in time.

In some cases, we still had to leave the ClearCache out of it (e.g. Echo), but other than those cases it feels much better! These are also available to the user so that if they have to call anything, then they still can.

PR #280, opened 2026-06-19
Update launch.json to include (No Debugger)

PaperPrototype:


PaperPrototype:
welp that didn't do anything, prowl still refuses to launch another instance during script reloading

PR #306, opened 2026-07-14
[Feature] SMAA 1x Anti-Aliasing

keithwill:
This PR adds **SMAA 1x** (Subpixel Morphological Anti-Aliasing) as a new post-process
image effect, alongside the existing FXAA and TAA options. It gives Prowl a spatial,
single-frame anti-aliaser that produces cleaner edges than FXAA while
avoiding the temporal ghosting that TAA can introduce.

## SMAA

SMAA is a well-established, GPU-friendly anti-aliasing technique that detects geometric
and shading edges in the final image and reconstructs a smoother result by re-blending
neighboring pixels according to precomputed coverage patterns. Compared to Prowl's
existing options:

- **vs. FXAA** — FXAA is cheap but tends to over-blur and soften fine detail and text.
  SMAA detects edges more precisely (with diagonal and corner handling) and produces
  crisper results with far less loss of sharpness.
- **vs. TAA** — TAA gives excellent coverage but relies on history reprojection, which
  brings ghosting, smearing on motion, and a dependence on jitter/motion vectors. SMAA
  1x is purely spatial and single-frame, so it has **no temporal artifacts** and no
  history buffers — a good fit for scenes where TAA's ghosting is objectionable or where
  the extra plumbing isn't wanted.

SMAA fills the gap between "cheap but blurry" (FXAA) and "sharp but temporal"
(TAA): a clean, self-contained spatial AA with predictable cost and no motion artifacts.

## How it was implemented

The effect is implemented as three chained fullscreen passes, driven from C# and
matching the reference SMAA pipeline exactly:

1. **Edge detection** — scene color → edges (luma-based).
2. **Blend weight calculation** — edges + the precomputed `AreaTex` / `SearchTex`
   lookup textures → per-pixel blend weights.
3. **Neighborhood blending** — scene color + blend weights → anti-aliased result.

Quality is set to the reference **HIGH** preset (16 search steps, 8 diagonal steps,
corner detection), with the edge-detection threshold exposed live as a tunable uniform.

### Conventions followed

**`ImageEffect` convention (mirrors `FXAAEffect`).**
`SMAAEffect` is a `sealed` `ImageEffect` and follows the same shape as the existing
effects:
- Lazily constructs its `Material` from `Shader.LoadDefault(DefaultShader.SMAA)` in
  `OnRenderEffect`.
- Sets uniforms, allocates pooled render targets via
  `RenderTexture.GetTemporaryRT`, drives the passes with `Graphics.GetCommandBuffer` +
  `cmd.Blit(...)`, submits, and releases the temporary RTs.
- Disposes the material in `OnDisable`.
- Registered in the `DefaultShader` enum as `SMAA` (right after `FXAA`), so it loads
  through the same default-asset path as every other built-in effect.

Like FXAA, it runs on LDR/perceptual color and is documented to be placed **late** in
`Camera.Effects` (after tone-mapping), and not stacked with FXAA or TAA. The intermediate
edges/weights buffers are kept at `RGBA8` (they're data, not color), while the final
blended output uses the scene color format.

**Reference implementation and attribution.**
The heavy lifting is the upstream reference SMAA from
[iryoku/smaa](https://github.com/iryoku/smaa) (MIT licensed). The reference
`SMAA.hlsl` is vendored **verbatim** (from the "SMAA Presets" section onward) as a Prowl
GLSL include at `Assets/Defaults/SMAA.glsl`, consumed via `#include "SMAA"` from
`SMAA.shader`. Each pass configures `SMAA_RT_METRICS`, selects the `SMAA_GLSL_4`
backend, sets the quality defines, and toggles `SMAA_INCLUDE_VS` / `SMAA_INCLUDE_PS`
before including — i.e. the standard way the reference is meant to be integrated, so the
upstream code is used unmodified.

- The original copyright notice (Jimenez, Echevarria, Masia, Navarro, Gutierrez, 2013)
  is retained at the top of `SMAA.glsl`, with a link to the upstream license. As
  permitted by the MIT license, only the long doc-comment header was trimmed; the
  copyright/notice is kept intact.
- The precomputed lookup-texture data (`AreaTex` / `SearchTex`) also originates from the
  same repository (`Textures/AreaTex.h`, `Textures/SearchTex.h`), and this is
  documented in `SMAALookupTextures`.

**Mechanism for including the LUT data.**
SMAA needs two precomputed lookup textures. Rather than ship them as PNGs, they're
included as **raw RGBA8 binary blobs** and uploaded directly to the GPU:

- The reference `RG8`/`R8` data is expanded into the R/G channels
  (`AreaTex = [R, G, 0, 255]`, `SearchTex = [R, 0, 0, 255]`) and stored as
  `Assets/Defaults/SMAAAreaTex.bin` (160×560) and `Assets/Defaults/SMAASearchTex.bin`
  (64×16).
- They're embedded through the existing `Assets\Defaults\**` embedded-resource glob and
  loaded at runtime via `EmbeddedResources.GetStream`, then uploaded with `SetData` into
  a `Color4b` `Texture2D` (Linear filtering, ClampToEdge). Each texture is cached as one
  shared GPU resource.
- Follows the existing precedent of `BRDFLutGenerator`, which loads its BRDF LUT from a
  raw `.brdf` blob for the same reason.

## Usage

Add the effect to a camera and place it late in the effect chain (after tone-mapping,
same as FXAA):

```csharp
camera.Effects.Add(new SMAAEffect { EdgeThreshold = 0.1f });
```

`EdgeThreshold` (default `0.1`, useful range ~`[0.05, 0.2]`) controls edge sensitivity —
lower catches more edges at higher cost. Don't stack SMAA with FXAA or TAA.

## Testing

Added `SMAAEffectTests` — headless smoke tests that verify:
- the four SMAA assets (shader, glsl, two LUT binaries) are embedded;
- `SMAA.shader` parses into the three expected passes and the `#include "SMAA"` is
  actually inlined (a reference-only function shows up in the compiled pass source);
- the AreaTex/SearchTex load at their expected dimensions (160×560 and 64×16).


keithwill:
AA Off
image

SMAA
image

michaelsakharov:
Looks good!, Thanks!

PR #311, opened 2026-07-22
Enable VS-Sync by default

PaperPrototype:
keep my PC from catching on fire


michaelsakharov:
We need a proper ingame API to control this and application target framerate. And we also need the Editor to relinquish control to game code in playmode. So if a game sets application.TargetFramerate = 30; it locks to 30, then exiting playmode should return to editor control. same with VSync, the game controls VSync when in playmode. That way if the editor locks to VSync on and TargetFramerate at like 60 or something, entering playmode disabled VSync and sets TargetFramerate to 0 (unlimited) by default. so the game gets a default state and can control it.

PR #317, opened 2026-07-23
Update launch.json

PaperPrototype:


PaperPrototype:
the debugger was stopping unecessarily, this fixes it

PR #323, opened 2026-07-24
Modify import and asset/scene references to use the actual names, rather than the EngineObject constructor default of New (TypeName)

Acissathar:
# Summary

Two small tweaks to how references display in the Editor. Currently, they always display as "New {TypeName}" or "New {TypeName} (Instance)" which can make it a little difficult to see at a glance what they are referencing. While clicking it does highlight / take to the right spot, that requires action from the user.

# Implementation

Import Result already tried to do this, but the problem was that the code never fired because Name was never null/empty. The EngineObject constructor always set it to be New {TypeName} no matter what.

Similarly, AssetRef and EngineObject just grabbed the Instance.Name, which was the EngineObject constructor default of New {TypeName}.

ImportResult just got rid of the dead guard, while in AssetRef and EngineObjectPropertyEditor I extracted out to a small helper that first checks if it is a component. If so, we use the GameObject's name and the Component's type in the display (this is a small change, previously it would always just be (Instance)). If a gameobject or asset, we use its Name (as now set properly from ImportResult). As a fallback, we use Instance still for the suffix (or None if instance is not true).

# Screenshots

Asset Reference pre-fix:
Screenshot 2026-07-24 160458

Asset Reference post-fix:
Screenshot 2026-07-24 160519

Scene Reference pre-fix:
Screenshot 2026-07-24 161827

Scene Reference post-fix:
Screenshot 2026-07-24 164306

# Notes

- This doesn't fix existing references (for assets. Scene Objects should be picked up). If an asset was created and assigned, it will need to be Reimported to be picked up in the inspector with the new name. Changing the name also appears to trigger this for what its worth.  Conveniently, the editor has a reimport all button at the top!

### DIFF
--- a/Prowl.Editor/AssetsDatabase/Importers/AssemblyDefinitionImporter.cs
+++ b/Prowl.Editor/AssetsDatabase/Importers/AssemblyDefinitionImporter.cs
@@ -35,7 +35,7 @@ public class AssemblyDefinitionImporter : AssetImporter
             ScriptAssemblyManager.RequestRecompile();
 
         // Marker asset so the inspector resolves the editor by type (data stays in the file).
-        ctx.SetMainAsset(new AssemblyDefinitionAsset { Name = Path.GetFileNameWithoutExtension(ctx.AbsolutePath) });
+        ctx.SetMainAsset(new AssemblyDefinitionAsset { Name = ctx.FileName });
         return true;
     }
 
--- a/Prowl.Editor/AssetsDatabase/Importers/AudioImporter.cs
+++ b/Prowl.Editor/AssetsDatabase/Importers/AudioImporter.cs
@@ -1,5 +1,3 @@
-using System.IO;
-
 using Prowl.Runtime;
 using Prowl.Runtime.Resources;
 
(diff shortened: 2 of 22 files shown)


michaelsakharov:
I think for assets, while it would be more code, I feel it makes more sense to have assigning names to the Main Assets to be the Importers job rather then SetMainAsset()
The Importers maybe can have a utility for it inside ImportContext, so something like this:
```cs
public string FileName => Path.GetFileNameWithoutExtension(AbsolutePath);
```
SetMainAsset instead of ever assigning a name, should probably throw if a Name is empty/whitespace, same with AddSubAsset in there a missing name literally breaks the asset. That way name control is 100% up to Importers.

Overall this is a good change though! I'll merge once thats tweaked!

michaelsakharov:
Oh also for Assets you show the Type Name?
Maybe "(Asset)" would be nicer? Similar to the "(Instance)" To make it clear what kind of reference it is?
What do you think?

Acissathar:
Sounds good, I will check it out!

Personally I'm partial to the type because then you can see at a glance exactly what kind of asset it is, like with the components. For example, material and input action map at a glance (granted these are AssetRef<>).

image

Unity gives them different icons (assets use their icon, gameobjects use a wireframe, transform is the gizmo, and components are just the script icon) for what its worth, but that's out of scope for me haha

If you'd prefer Asset though to match Instance I can make the swap, I think the Name is the biggest with regards to clarity.

michaelsakharov:
lets go with the name then, im need to make a proper system for asset icons, i'll revisit the naming here once i do, Thanks!

Acissathar:
Meshes/textures/etc.  types with hand-written serialization weren't persisting Name, so it reverted on reload. I added the header write/read (via SerializeHeader) to the ones that were missing it, plus importer version bumps so existing caches regenerate.

PR #324, opened 2026-07-25
Improved rename flow and added auto-expanding/better context menu flow for hierarchy panel

xZekro51:
Just as stated in the title, the rename flow for GameObjects in the hierarchy was improved by using an optimally sized text field, as well as auto-expanding objects when a child is being created.


PaperPrototype:
I want this!

PR #325, opened 2026-07-26
Fix: Cameras that are a child object are being rendered multiple times in Game View

Acissathar:
I wired up a stats panel and noticed it said two cameras and two directional lights were being rendered despite me only having one. Turned out its because my camera is a child object of another, and digging found this is unique to the Game View. 

Scene.Render already accounts for this exact scenario (which is also what's used for actual builds) by calling Distinct on the fetch for explicitly this reason, but Game View did not replicate it. In theory, this scales for every ancestor the camera has.

Just copied over the tweak and the same comment explaining it.


michaelsakharov:
Oh! Good find!

PR #327, opened 2026-07-29
Add Align With View to the hierarchy context menu

Porphyri0n:
Adds Unity's **Align With View** to the GameObject context menu in the Hierarchy: it moves the selected object(s) to the scene view camera's position *and* rotation, so the object ends up looking exactly where the view looks. Handy for placing cameras and spotlights — fly the view to the shot you want, then align.

### Why "Move to View" changed too

The existing `Move to View` already applied position **and** rotation, which is exactly what Align With View means — so adding Align With View on its own would have produced two menu entries doing the same thing, and nothing doing Unity's actual Move to View.

`Move to View` now moves objects to the centre of the view and leaves rotation untouched, matching Unity. The two commands are distinct:

| Command | Position | Rotation |
| --- | --- | --- |
| Move to View | centre of the view | unchanged |
| **Align With View** (new) | camera position | camera rotation |
| Move View To | — | moves the camera to the object |

### Other fixes in the touched code

Both commands now:

- **Register a single undo step.** `Move to View` previously recorded no undo at all, so a misclick permanently lost the object's old transform.
- **Write world-space rotation** via `SetPositionAndRotation` instead of assigning the camera's *world* angles to `LocalEulerAngles`, which produced the wrong orientation for a child of a rotated parent.
- **Apply to the top-level selection only** (`ExcludeNestedSelections`), matching Unity's `Selection.transforms`. A child whose parent is also selected already travels with the parent, so writing its world pose separately fought the parent's write and made the outcome depend on iteration order.

`EditorCamera` gains two members: `Rotation` (world rotation from the current yaw/pitch) and `ViewFocusPoint` (the point the view is centred on). `Move to View` drops objects at the focus point rather than at the camera itself, where the near plane would clip them out of sight.

`hierarchy.align_with_view` is added to all 12 locale files, following each file's existing convention (fr/pl/tr are ASCII-only).

### Testing

Built clean and verified in the editor: align on a root object and on a child of a rotated parent, undo/redo for both commands, multi-selection including a parent + child pair, and that `Move to View` no longer disturbs rotation.

### DIFF
--- a/Prowl.Editor/GUI/Panels/HierarchyPanel.cs
+++ b/Prowl.Editor/GUI/Panels/HierarchyPanel.cs
@@ -786,19 +786,34 @@ public class HierarchyPanel : DockPanel, IScriptReloadCleanup
 
             builder.Separator();
 
-            // Move to View / Move View To
+            // Move to View / Align With View / Move View To
             var cam = SceneViewPanel.ActiveCamera;
             if (cam != null)
             {
+                // Pose writes go to the top-level selection only: a GameObject whose ancestor is
+                // also selected already travels with that ancestor, so writing its world pose
+                // separately would fight the parent's write (order-dependent) and bloat the undo step.
+                var poseTargets = ExcludeNestedSelections(selectedGOs);
+
+                // Position only, at the centre of the view - rotation is deliberately untouched.
                 builder.Item(Loc.Get("hierarchy.move_to_view"), () =>
                 {
-                    foreach (var go in selectedGOs)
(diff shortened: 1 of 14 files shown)


Porphyri0n:
The 'Align with View' and 'Move to View' commands were the ones I used most frequently while editing scenes, and I have truly felt their absence. I would like to request that these commands be added to the GameObject dropdown menu in the top toolbar as well. Please provide your feedback on this matter.

PR #332, opened 2026-08-01
Fix periodic frame spikes caused by per-frame LOH allocations in PaperRenderer

Acissathar:
# Summary

While trying to help flatten my frametime in my project, I found that any Paper-rendered UI (including the editor itself) suffered periodic frame spikes on a metronomic ~0.3 s beat of roughly 10–35 ms frames against a 5.6 ms baseline. This was reproduced in both the editor and standalone builds.

After some digging and trimming down my own allocations from the project itself, we were ablet o land on the cause being the PaperRenderer.RenderCalls rebuilding its vertex/index uploads from scratch every frame per canvas.

This quickly pushed us past the 85KB LOH threshold and drove GC, forever.

# Fix

Fortunately its a pretty easy fix, we just allocate the arrays once and then re-use them instead of allocating every frame.

Stats from my project directly:

| Metric | Before | After |
| --- | --- | --- |
| Spikes detected (>1.8× avg) | 693 | 1 (a benign startup gen1) |
| Frames over 16.67 ms budget | 302 / 26,566 (1.1%) | **0 / 16,881** |
| 120-frame window min/max | 3.4 / 32.5 ms | 5.0 / **6.0** ms |
| Main-thread alloc | 2.3 MB/frame | 1.2 MB/frame |
| GC pause | 0.126 ms/frame | 0.036 ms/frame |
| gen2 as share of gen0 | ~62% | ~2% |

Note that this doesn't affect the average frame time, that's still roughly unchanged but more of a pacing fix and removing a noticeable consistent micro stutter.

### DIFF
--- a/Prowl.Runtime/GUI/PaperRenderer.cs
+++ b/Prowl.Runtime/GUI/PaperRenderer.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.InteropServices;
 
 using Prowl.Quill;
@@ -33,6 +32,10 @@ public class PaperRenderer : ICanvasRenderer
     private const int MaxBlurLevels = 6;
     private Resources.Material _blurMat;
     private readonly List<RenderTexture> _tempBlurRTs = new();
+
+    // Reused upload scratch that grows to the UI's high-water mark once and is then allocation-free.
+    private Vertex[] _vertexScratch = Array.Empty<Vertex>();
+    private uint[] _indexScratch = Array.Empty<uint>();
 
     public void Initialize(int width, int height)
     {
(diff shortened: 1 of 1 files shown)


michaelsakharov:
I just made a similiar fix to other quill backends actually! including improvements to the quill API to reduce allocations internally as well, I'll update prowl to it, see if that fixes this issue before i merge this.

Acissathar:
> I just made a similar fix to other quill backends actually! including improvements to the quill API to reduce allocations internally as well, I'll update prowl to it, see if that fixes this issue before i merge this.

Fixed the issue *and* I can see the extra performance improvements. Spike is gone *and* avg frame time was cut in half (~6ms to ~3ms)

PR #336, opened 2026-08-05
Fixed DontDestroyOnLoad, added compile datetime check for project assemblies

xZekro51:
DontDestroyOnLoad previously kept objects alive even after exiting play mode, causing clashes with the pre-existing objects in the scene. This brought up an issue with Assemblies as well, where if the last time they were compiled was before the current engine version, they could've referenced missing parts of the API, causing errors.

Note: I did not double-check the tests, but they looked sound to me. Are they ok?

### DIFF
--- a/Prowl.Editor.Test/ScriptCompilationTests.cs
+++ b/Prowl.Editor.Test/ScriptCompilationTests.cs
@@ -185,6 +185,25 @@ public class ScriptCompilationTests : EditorTestHarness
         finally { TryDeleteDir(parent); }
     }
 
+    // Script files are the only input the recompile rule watches, so an engine rebuilt underneath an
+    // unchanged project leaves assemblies bound to an API that may no longer exist - surfacing much
+    // later as a MissingMethodException from whichever call site happened to run first.
+    [Fact]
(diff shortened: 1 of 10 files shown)


michaelsakharov:
Instead of trying to guess if the assembly is outdated by looking at last modified dates on files

It might make more sense to just always recompile scripts on project load?
Its only a couple seconds, and it guarantees everything is up to date if something changes while the editor is off or something, and it also immediately throws compile errors into the console to let the user know about them straight away

xZekro51:
> Instead of trying to guess if the assembly is outdated by looking at last modified dates on files
> 
> It might make more sense to just always recompile scripts on project load? Its only a couple seconds, and it guarantees everything is up to date if something changes while the editor is off or something, and it also immediately throws compile errors into the console to let the user know about them straight away

I implemented your suggested behaviour with the latest commit, I was initially a bit conflicted on how to do it and went with the date as it felt a bit wasteful to always recompile on project load, but in the end it doesn't take _that long_ to recompile it, so it made sense.

I added a RequestRecompile after each LoadAssembly, the only site I left untouched was `ScriptAssemblyManager.InitialLoad` as that is used by the normal recompile process, is that correct?

PR #345, opened 2026-08-05
`f` key focuses Game Canvas in scene view correctly

PaperPrototype:
https://github.com/user-attachments/assets/3efadb86-3818-40e0-a44c-0a05517de2e8

clicking `f` to focus the Game Canvas now works correctly. Though by default you have to zoom in. I'm not sure if that is because max render distance is clipping it or not? @michaelsakharov any idea why it is not visible?

### DIFF
--- a/Prowl.Editor/GUI/SceneView/EditorCamera.cs
+++ b/Prowl.Editor/GUI/SceneView/EditorCamera.cs
@@ -518,6 +518,26 @@ public class EditorCamera
             max = new Float3(MathF.Max(max.X, wb.Max.X), MathF.Max(max.Y, wb.Max.Y), MathF.Max(max.Z, wb.Max.Z));
             any = true;
         }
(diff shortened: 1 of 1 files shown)


michaelsakharov:
The Editor has a Far Clip plane of I think 1000? which is reasonable, the issue is your focus is just placing the camera a bit to far back. Im not sure but maybe Unity adjusts far plane depending on last focused object? Im too tired to test that right now, i'll investigate how unity handles this tomorrow.

sinnwrig:
Unity changes the min/max clip of the scene view camera to match the object selected when dynamic clipping is on. 
Defaults are 0.03/10000 iirc
you could probably just change the scene camera near/far to be something like this, and have the scale be something like the distance of the scene view camera to the pivot of the selected object

```
near = max(minNear, scale * nearFactor)
far  = max(near + margin, scale * farFactor)
```

michaelsakharov:
This has been done properly now

PR #354, opened 2026-08-16
Fix typed characters never reaching the Game View's Paper

Acissathar:
Text fields drawn from MonoBehaviour.OnGui accepted Backspace/Delete and arrow keys but no printable characters at all while in play mode in the editor.

The editor's Paper is pumped during Update, before the clear, so inspector fields worked. The Game View's Paper is pumped during Render, after it, so PaperInputBridge's foreach (char ch in Input.InputString) always iterated an empty string. Editing keys were unaffected because they come from Input.GetKeyDown/GetKeyUp, backed by dictionaries that UpdateKeyStates() refreshes at the end of LateUpdate and which stay valid for the rest of the frame.

The fix moves the clear to a new DefaultInputHandler.BeginFrame() called immediately before DoEvents(), so a frame's characters survive from the events that produce them through to PostRender. That matches what the field's own comment already claimed (that any number of consumers can read it each frame) which previously only held for Update-phase consumers.


PaperPrototype:
Ayy it's only 2 files

PR #355, opened 2026-08-18
Add configurable shadow focus target for directional cascades

keithwill:
Directional-light shadow cascades were always centered on the rendering camera's position, which assumes a first-person view. In a third-person game the player character is a distant focal point, so the character lands in a low-resolution far cascade (or falls out of the cascade range entirely) while the resolution is spent on empty ground in front of the camera.

Add an optional per-camera Camera.ShadowFocus transform that cascades center on instead. It falls back to the camera's own position when unset or when the target's GameObject has been destroyed, so existing scenes and the editor's scene view (which builds a fresh temporary Camera) are unaffected.

The CPU side places cascades around the focus point in DirectionalLight.GetShadowMatrix, while two shader sites select a cascade by distance: SampleDirectionalShadow in Lighting.glsl and VolDirShadow in VolumetricFog.shader. Both have to move together or selection disagrees with placement and cascade seams appear, so the resolved point is uploaded as a new _ShadowFocusPos uniform and both sites measure from it. The uniform is written before the no-directional-light early-out so it stays fresh every frame. Camera-relative values (view direction, fog distance, specular) keep using _WorldSpaceCameraPos.

The point is resolved once per render into CameraSnapshot, so mutating the target mid-render can't desync placement from selection. Closest-N local shadow caster selection in SceneLightSystem.Reconcile uses the same point.

Known limitation, unchanged by this: the cascade ortho depth range is a fixed +/- cascadeDistance * 0.5 slab around the snapped focus point rather than a fit to the scene's casters, so tall occluders toward the light can clip out of the map. This relocates that behaviour with the focus point; it does not fix it.

Tests cover the resolver's fallbacks, that it tracks a moving target, that the cross-object Transform reference survives a scene save/load via Echo's intra-graph rewiring, that a cascade lands centered on its focus point to within half a texel, and that sub-texel movement of the focus produces an identical view matrix (the texel snapping that keeps shadow edges from shimmering).

The BananaMan sample now demonstrates it: a player stand-in with props sits 25 units out on a large ground plane, the camera is third-person, and F toggles the focus between the player and the camera.

### DIFF
--- a/Prowl.Runtime.Test/DirectionalLightShadowMatrixTests.cs
+++ b/Prowl.Runtime.Test/DirectionalLightShadowMatrixTests.cs
@@ -1,0 +1,81 @@
+// This file is part of the Prowl Game Engine
+// Licensed under the MIT License. See the LICENSE file in the project root for details.
+
+using System;
+
+using Prowl.Vector;
+
+using Xunit;
+
+namespace Prowl.Runtime.Test;
+
+/// <summary>
+/// Pure-math tests (no GPU) for <see cref="DirectionalLight.GetShadowMatrix"/>: a cascade lands
+/// centered on the focus point it is handed - the rendering camera, or the camera's
+/// <see cref="Camera.ShadowFocus"/> target - and the light-space texel snapping that keeps shadow
+/// edges from shimmering survives sub-texel movement of that point.
+/// </summary>
+public class DirectionalLightShadowMatrixTests : RuntimeTestBase
+{
+    private const int Resolution = 2048;
+    private const float CascadeDistance = 35f;
+    private const float TexelSize = (CascadeDistance * 2f) / Resolution;
+
+    /// <summary>Creates an angled directional light, so the light-space axes are nothing like the
+    /// world axes and a mistake in the basis math can't accidentally cancel out.</summary>
+    private DirectionalLight CreateAngledLight()
+    {
+        GameObject go = CreateGameObject("Directional Light");
+        go.Transform.LocalEulerAngles = new Float3(-50f, 30f, 0f);
+        return go.AddComponent<DirectionalLight>();
+    }
+
+    /// <summary>The orthonormal light-space basis GetShadowMatrix builds internally.</summary>
+    private static (Float3 right, Float3 up, Float3 forward) LightBasis(DirectionalLight light)
+    {
+        Float3 forward = -light.Transform.Forward;
+        Float3 up = Float3.Normalize(light.Transform.Up);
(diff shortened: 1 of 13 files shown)


Fabrizio86 (approved):

PaperPrototype:
Can you add screenshots of it working (I'm just curious and too lazy to run it myself atm)

Prowl.Runtime.Test/DirectionalLightShadowMatrixTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        // boundary and legitimately land on the next grid point.
+        Float3 focus = (right * (14f * TexelSize)) + (up * (-9f * TexelSize)) + (forward * 6.5f);
+        Float3 nudged = focus + (right * (TexelSize * 0.2f));
+
+        light.GetShadowMatrix(focus, Resolution, CascadeDistance, out Float4x4 view, out _);
+        light.GetShadowMatrix(nudged, Resolution, CascadeDistance, out Float4x4 nudgedView, out _);
+
+        // Identical, not merely close: a shadow map that slides by a fraction of a texel each frame
+        // is what makes shadow edges crawl, and this is exactly the case a moving player hits.
+        Assert.Equal(view.ToArray(), nudgedView.ToArray());
```
Fabrizio86: Exact float equality here is brittle. GetShadowMatrix snaps X and Y but not Z, and Dot(nudged, forward) picks up 0.2 * texel * Dot(right, forward), which is ~1e-8 rather than zero since the basis comes out of normalize/cross in float. That noise flows into snappedPosition and the view matrix, so this passes only if the float error happens to cancel. Suggest asserting the X/Y translation is bit-identical (that's the shimmer guarantee) and Z within an epsilon, or snapping Z as well if the ortho slab can tolerate it.

keithwill:
@PaperPrototype 

> Can you add screenshots of it working (I'm just curious and too lazy to run it myself atm)

Here is a very contrived ad-hoc example of it functioning (not necessarily one meant to make it look good or useful). The camera is ~1000m from the focal point and there are shadows 400m out from the focal point. The yellow circle I added just for illustration.

image

To do this centered on the camera would require me to cast shadows out much farther. What I was doing before the change (forgive the quick MS paint jank):
image

PaperPrototype:
I understand the concept, was just curious how it looked with it on vs off. Anyway, it's merged now :P

keithwill:
@PaperPrototype 

Gotcha. This is from the Banana man sample.

Off:
image

On:
image

PR #359, opened 2026-08-19
Macro Texturing

PaperPrototype:
Draws a scaled up texture for far away terrain to hide tiling. Seamlessly blends between scaled and unscaled nearly invisible unless you make the fade distance really small.

Before (notice the texture tiling)
Screenshot 2026-08-18 at 6 21 06 PM

After (notice how far away terrain looks so nice?)
Screenshot 2026-08-18 at 11 14 14 PM

This also adds 2 more sections to the terrain settings to control distance tiling
Screenshot 2026-08-19 at 3 42 28 AM


PaperPrototype:
Note: this PR is not meant to be merged, only for demonstration purposes.

Terrain currently lacks any sort of Heightmap texture support, so the Layer Blending fakes it using normals. Once proper heightmap texture support is added only then can layer blending can be properly implemented using that.

PR #363, opened 2026-08-23
Pass VSync toggle during rendering rather than only at startup

Acissathar:
The swap interval is pushed once, at Start(). InternalWindow.VSync is the property Silk's own render loop would consult — and Prowl doesn't use that loop. So the setter updates a value nobody reads, and the toggle can never do anything at runtime. It only works via InitWindow's options because Start() reads it once afterwards.

This change makes it so that the setter now also stores it as a pending swap, which is read during the render loop as it needs the context.

I know there was mention on the Editor VSync issue (which this also covers in addition to builds) about looking into a more robust solution (framerate caps, etc.) so just wanted to at least throw this up should someone else want it if this isn't something that you want merged in lieu of the other solution.

### DIFF
--- a/Prowl.Runtime/Graphics/Graphics.cs
+++ b/Prowl.Runtime/Graphics/Graphics.cs
@@ -232,6 +232,12 @@ public static unsafe class Graphics
 
                 if (job.IsFrameEnd)
                 {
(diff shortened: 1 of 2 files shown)


michaelsakharov:
Vsync and FPS Limiting is properly handled in the Editor and Playmode now

PR #366, opened 2026-08-27
Fix NRE when deleting a texture or sprite in the Editor

Acissathar:
Trying to delete a sprite/texture in the Editor usually gives me a null reference that crashes the editor (it does still delete). Issue was they were using old null check instead of Valid/NotValid


michaelsakharov:
== and !=
Shooouuld actually be okay to use, the Analyzer should catch anything at fault.
I think the real issue you caught here, was the old code was checking the AssetRef for null, not the asset itself.

Nice find! Thanks!

PR #367, opened 2026-08-27
Add support for drag and drop asset importing

Acissathar:
Currently to add assets to Prowl (models, textures, etc.) you have to copy in Explorer and then Prowl will import. This hooks in the FileDrop event to import the given asset to Prowl.

Supports dragging to a specific folder and importing copies of something that already exists by adding (x) to the name.

https://github.com/user-attachments/assets/aafacabc-2734-4737-9957-66095a9867d4


michaelsakharov:
This is Awesome! Its been on my todo list for ages now
Thanks!

Only thing its missing is dragging into the Scene view should Import, and we need to find a way for the IMporters themselves to be included to also bring Textures and helper files along, like if i drag a gltf into the scene/assets, the Importers should somehow trigger to search for all relevant files around the gltf, and import all those too in the same structure, otherwise the gltf doesnt work without them.

But this is a huge stepping stone toward that, Thanks!

PR #368, opened 2026-08-28
Screenshot support

draxxris:
Support taking screenshots in engine itself. (Used for automation tests in downstream project)

```
- Game: add virtual AfterGui(Scene?) called after _paper.EndFrame() before backbuffer swap. Presentation hosts use this to capture the rendered frame on the render thread (for automation screenshot).

- Graphics: restore ReadPixels + CommandBuffer/Executor/Opcode for framebuffer readback (removed upstream).

- Prowl.Runtime: bump Magick.NET-Q8-AnyCPU 14.13.0 -> 14.16.0 to fix known vulnerabilities.
```

### DIFF
--- a/Prowl.Runtime/Game.cs
+++ b/Prowl.Runtime/Game.cs
@@ -172,6 +172,11 @@ public abstract class Game
 
                 _paper.EndFrame();
 
+                // Give presentation hosts a point after all scene and Paper commands have been
+                // submitted but before the backbuffer is swapped. Development screenshot capture
+                // uses this hook to read the actual rendered client on the render thread.
+                AfterGui(currentScene);
+
                 // === End Graphics ===
 
                 RenderTexture.UpdatePool();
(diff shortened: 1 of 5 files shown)


michaelsakharov:
It might be nicer to have a `Screenshot()` method rather then a `ReadPixels(int, int, uint, uint, byte[])`?

Because `Graphics.ReadPixels` could mean a few things.
And raw byte data isn't as helpful as like returning a Texture2D so
`public static Texture2D Screenshot()`

Then they get back a normal Texture2D and can do whatever they want with it.

draxxris:
Added Screenshot() that calls ReadPixels()

Why not replace ReadPixels() entirely? Eventually plan to add StartVideoRecord() and StopVideoRecord() to create videos.

michaelsakharov:
Oh thats what I meant, remove ReadPixels in favor for a nicer and easier to use API Screenshot(), rather then adding, since Graphics.ReadPixels is still quite vague.

draxxris:
That is done now.

PR #375, opened 2026-09-12
Modified icons invocations to standardize them

xZekro51:
As the title states, the purpose of this is to standardize icon calls in Hierarchy and NewScriptDialog so that they can be modified more easily in the future if necessary. Furthermore, this makes the size of icons in the Hierarchy and in the NewScriptDialog a bit larger so that they're easier on the eyes.

Size increased
image

Original size
image

This also opens a question: do we want to make this larger in scope? I do believe there are more areas of interest that use Icon which could be standardized (i.e. the Component header row in the GameObjectInspector), should this change extend into those areas as well? Let me know and I can make the changes.

### DIFF
--- a/Prowl.Editor/GUI/Panels/HierarchyPanel.cs
+++ b/Prowl.Editor/GUI/Panels/HierarchyPanel.cs
@@ -366,8 +366,8 @@ public class HierarchyPanel : DockPanel
                         var (goIcon, goColor) = GetGoStyle(go);
                         if (!go.EnabledInHierarchy) goColor = Color.FromArgb(120, goColor);
                         paper.Box($"hier_ico_{goId}")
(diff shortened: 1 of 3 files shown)


michaelsakharov:
I think so yeah, like standardize it everywhere.
Theres actually quite a bit of the Editor UI that could be standardized, like some buttons or logic not using Origami, or some things not setting up theme correctly.

I Wrote the Editor, then improved Paper, then made Origami, then rewrite Editor icons. So like, theres a lot of UI scattered around using legacy methods like those buttons you replaced

michaelsakharov:
Would you like me to merge this now? or keep it open?

xZekro51:
I'd say keep this open so that I can fix all the other sites first- Then it'd be just one larger PR to review 👍

Issue #52, opened 2023-11-25
[Editor] Hotkey/Shortcut system


michaelsakharov:
Shortcuts like Ctrl+D to duplicate should probably be merged into the existing input system.
It might work something like

```cs
if (Input.ShortcutPressed("Duplicate"))
{
    // duplicate here
}
```

But before this, we need Input settings for Projects

michaelsakharov:
Found this useful list of all the shortcuts Unity has:
https://docs.unity3d.com/2018.1/Documentation/uploads/Main/Unity_HotKeys_Win.pdf

I guess we should try and recreate as many of those as possible

PR #380, opened 2026-09-13
Navigation: tile-cache-backed navmesh, crowd agents, and seven capabilities beyond Unity

Fabrizio86:
# Navigation: tile-cache-backed navmesh, crowd agents, and seven capabilities beyond Unity

## Summary

Replaces the navigation subsystem with one built on the same DotRecast foundation but a different core
architecture: every surface bakes as a tile cache (never a frozen finished-tile mesh), which is what makes
runtime carving, partial retiling, and tile streaming possible everywhere, unconditionally, with nothing to
configure per surface. On top of that foundation this branch adds seven capabilities that neither Unity's
`com.unity.ai.navigation` package nor #335 have: dynamic decaying cost overlays, live corridor lookahead,
automatic jump/drop link generation at bake time, shared flow fields for crowd-scale movement, scored
tactical position queries, hierarchical (coarse-then-detailed) pathfinding for large navmeshes, and a
runtime tile streaming API. Full design rationale, API shape, and known limitations for all of it are in
`docs/navigation/design.md`, committed with the branch.

This is not a patch on top of #335; it is a from-scratch rebuild, which is why the mapping below is mostly
architectural rather than line-level. The whole subsystem lands as a single commit, with a full account
of the decisions, trade-offs, and each capability's own rationale in its own body.

## Capability list

| # | Capability | Neither Unity nor #335 have this |
|---|---|---|
| 1 | Dynamic, decaying cost layers per agent type (`NavMeshCostLayer`) | Yes - Unity's `SetAreaCost` is bake-time only |
| 2 | Live corridor lookahead (`UpcomingCorners`, `NextCornerDistance`, `NextTurnAngle`, `IsApproachingLink`) | Yes |
| 3 | Automatic jump/drop link generation at bake, scoped per agent type | Partial - Unity has a project-wide version; this one is per agent type and stores output on the baked asset rather than scene objects |
| 4 | Shared flow fields for many agents converging on one goal (`NavMeshFlowField`) | Yes |
| 5 | Scored tactical position queries (`QueryTacticalPositions`, `ITacticalScorer`) | Yes |
| 6 | Hierarchical (coarse tile graph, then restricted detailed search) pathfinding | Yes |
| 7 | Runtime tile streaming (`UnloadTiles`/`LoadTiles`) independent of rebake | Yes |

Each has its own section in `docs/navigation/design.md`, including a documented non-goal or limitation
where one exists (flow-field resolution being per-polygon, not per-voxel, and others listed there).
Cost-weighted search was initially believed to have a limitation of its own - the search never visibly
preferring a cheaper route - until decompiling Detour's own A* loop showed the cost formula was correct,
and the actual cause was the test fixture's own coarse bake (a small open area collapsing into one or two
polygons, leaving nothing for a cost to route around regardless of magnitude). Rebuilding the test with a
finer tile size confirmed the mechanism works as designed: the route bends around a costed region and the
direct route returns once the cost is cleared.

## How this addresses #335's maintainer review

#335 collected 30 review comments, mostly from `michaelsakharov`, covering both architecture-level concerns
and specific bugs. Below is how this branch's own design and architecture relate to each substantive point.
Several are moot because the component or pattern they were about no longer exists in this branch; those
are marked accordingly rather than claimed as fixes.

| #335 review point | This branch |
|---|---|
| PR should have been split into 3 (unrelated attribute/inspector work bundled with navigation) | Not applicable - this branch is scoped to navigation only, nothing unrelated is bundled |
| `NavMeshWorld.DefaultQueryExtents` is a public mutable `Float3` read from worker threads without synchronization | Not applicable - there is no `NavMeshWorld`; per-agent-type state lives on `NavMeshSystem`/`NavMeshQuery`, and cross-thread reads go through the query's own `ReaderWriterLockSlim`, taken for every structural mutation and documented on the class |
| "An agent with no neighbours has nothing to avoid" is wrong; boundary segments are still fed to avoidance | Not directly verified against DotRecast's own crowd internals; not re-asserted anywhere in this branch's own docs |
| `NavMeshAgent.Move()` tears down and rebuilds the crowd agent every call; `Warp` doesn't clear arrival state and writes an unsnapped position | This branch has no `Move()`; `Warp` sets the transform, drops and rejoins the crowd once (a genuine full reset, documented as such), and is explicitly scoped to teleport-style use, not per-frame driving |
| `NavigationSettings.CreateDefaultNames`/`CreateDefaultCosts` read live edited state, so "reset to defaults" doesn't reset | Not applicable - this branch has no `NavigationSettings` editor asset; `NavMeshAreas`/`NavMeshAgentTypes` reset via `CreateDefault()`, which only ever references compile-time constants, never live state |
| Unclamped tile rebuild loop, measured at 1.9s for 100k tiles, holding the write lock the whole time | `RebuildTiles`/`RebuildTilesAsync` are bounded to the tiles overlapping the caller's own region from the start (this is item 7's own API), and the async form splits geometry collection from the actual build specifically so the write lock is only held for the bounded apply step |
| `NavMeshLink` rebuild/catch-up walks the whole scene, then walks it again per surface, on every move and on `OnDisable` | This branch's `NavMeshLink` (`AutoUpdatePosition`/`AutoRebuild`, task #52) was written against this concern directly; not re-benchmarked at the same 30-link scale #335's review measured, so this is a design intent, not a re-verified number |
| `MaxTileCrossingConnections = 4` works around a DotRecast link-allocation bug that should be fixed upstream | This branch depends on `Prowl.Recast`, the project's own NuGet package (not raw DotRecast), which is exactly the fix the reviewer asked for in principle; whether this specific allocation guard is what changed in that package was not independently re-verified from this branch |
| Editor-mode navigation is inert (no `[ExecuteAlways]`), and a test that claims to cover it is broken by `RuntimeTestBase` forcing `IsPlaying = true` | Not independently re-verified; out of scope for this branch's own test additions, which are runtime-only like the rest of the suite |
| `NavMeshRasterizer`'s `[ThreadStatic]` scratch is never released, leaking memory | This branch's rasterizer is `RecastNavMeshBuilder`, a different implementation; not independently audited for the same pattern |
| Carve throughput hard-capped at one tile per frame, no way to tune it | `NavMeshQuery.TickTileCache` pumps `_tileCache.Update` with a named `TileCacheUpdateBudget` constant (currently 4) rather than an implicit single-tile pop, directly answering "there is no `MaxTileUpdatesPerFrame` anywhere" |
| `NavMeshChanged` fires every frame of every carve; agents aren't gated the way links/obstacles are | Not applicable - this branch has no `NavMeshChanged` event; an agent's own corridor is Detour's live, continuously replanned state, read directly (see item 2's lookahead API) rather than driven by a coarse changed-event |
| `ReaderWriterLockSlim` leak - `NavMeshInstance` isn't `IDisposable`, rebakes leak wait handles | Not applicable in the same shape - this branch's lock lives on `NavMeshQuery`, one per agent type, replaced (not accumulated) on `RebuildQuery`; disposal was not independently audited |
| `SetPath` silently ignores the given path and re-plans from its last corner | Not applicable - this branch has no `SetPath`; `FollowFlowField` (item 4) is the alternative to `SetDestination` this branch offers, and it does not silently substitute a different target |
| `IsPathStale` duplicates `PathStatus`'s own partial-path information | Not applicable - this branch has no `IsPathStale` |
| `NavMeshHit.Mask` never populated by `Raycast` | Not applicable - this branch has no `NavMeshHit` return type |
| `FindClosestEdge` hardcodes a 100-unit search radius | This branch's `FindClosestEdge` takes `maxRadius` as a caller-supplied parameter directly, not a hardcoded constant |
| `NavMeshLink.LinkId` is minted at runtime via `Random.Shared`, mutating scene state on first enable; `s_liveLinks` is an unsynchronized process-wide static | Not applicable in the same shape - this branch's link identity is the component instance itself, not a separately minted random id or a parallel static registry |
| `NavMeshData` assets are mutated in place, desyncing what the asset database handed out from the real asset | Still true of this branch's own design: `RebuildTiles`/tile streaming's `LoadTiles` write into the baked `NavMesh` asset's own `TileCacheData.Layers` in place, the same trade-off, not eliminated. Recorded as a design choice (mutation is what lets a reload skip re-baking) rather than re-litigated as a bug here |
| `NavMeshObstacle` bypasses the tile-cache mutation contract, calling `AddObstacle`/`RemoveObstacle` directly instead of going through a documented gate | This branch centralizes every tile-cache mutation inside `NavMeshQuery` itself (`AddBoxObstacle`, `AddCylinderObstacle`, `RemoveObstacle`, `TickTileCache`, all taking the query's own write lock); there is no separate side channel for `NavMeshObstacle` to bypass |
| `MaxPolyPath`/`MaxStraightPath` are private constants with no way to raise them for a long path | Not independently re-verified; this branch's `FindPath` does not surface a caller-configurable path-length ceiling either |
| `ProwlInputGeomProvider`'s area merge always lets a walkable span win over a nearby not-walkable one | This branch's geometry collection path is different (`NavMeshGeometryCollector`/`RecastNavMeshBuilder`); not independently re-verified against the same merge case |
| `NavMeshGeometryCollector` doc claims a mesh-collider triangle-extraction optimization that isn't actually implemented | Not independently re-verified against this branch's own collector |
| `NavMeshAreas`/`NavMeshAgentTypes` are mutable process-wide statics read from worker threads with no synchronization | Fixed. `NavMeshQuery.CreateFilter` (reachable from a worker thread on every masked query call, not just at construction) rebuilds its area-cost snapshot from `NavMeshAreas.GetAreaCost` on every call, so this was independently confirmed reachable, not just theoretical. Both tables now publish a whole new array on every write instead of mutating the live one's structure or a shared slot in place - `NavMeshAgentTypes` in particular no longer uses a plain `List<T>` mutated by `Add`/`Remove` while `Types` is enumerated elsewhere, which reproduced an intermittent exception under the added concurrency test before the fix (1 failure in 5 runs) and is stable across repeated runs after it. `NavMeshAgentTypeInfo`'s own fields remain directly mutable by reference (an existing, tested contract - `GetById(id).Name = ...` is expected to be visible on the next `GetById(id)`), so a caller holding a reference and racing a field write against a worker-thread read of that same instance is still possible in principle; nothing in this branch currently reaches that path from a worker thread, unlike the area-cost snapshot case, which was the one independently reproduced |
| `NavMeshBuildSettings.Clone()`/`NavMeshAgentType.Clone()` via `MemberwiseClone` is a redundant, fragile API | Not applicable - this branch has no `Clone()` method on the equivalent settings types |
| `.navmesh` assets are serialized as Echo text, making a real level's compressed tile data large and undiffable | Not independently re-verified; this branch's `NavMesh` asset serialization was not changed as part of this work |

## Unity comparison

Covered in full, with every claim sourced to Unity's own published `com.unity.ai.navigation` documentation,
in `docs/navigation/design.md` under "Comparison with Unity's NavMeshComponents". Summary: parity on
`NavMeshObstacle` carve modes, geometry collection scope/source, `NavMeshLink`, and area costs; this
branch's own seven capabilities above are each confirmed absent from Unity's documented feature set.

## Test counts

- 1011 tests passing in `Prowl.Runtime.Test`, all three builds (`Prowl.Runtime`, `Prowl.Editor`,
  `Prowl.Player.Desktop`) succeeding with `-tl:off`.
- 135 of those tests are navigation-specific, across 22 test files under `Prowl.Runtime.Test/NavMesh*.cs`.
- Tile streaming added 3 tests: unload/reload with a crowd slot held constant through the whole round
  trip, `LoadTiles` as a no-op when nothing was ever unloaded, and `RebuildTilesAsync` applying the same
  change as its synchronous counterpart.
- The registry thread-safety fix added 2 concurrent read/write tests, one per registry; both reproduced
  the pre-fix bug directly (an intermittent exception under `NavMeshAgentTypes`' old `List<T>`-based
  storage, 1 failure in 5 runs) before confirming the fix is stable across repeated runs.
- The cost-layer fix replaced a test that only checked a filter received the right value with one that
  checks the actual route: a path bends around a costed region and returns to the exact original route
  once the cost is cleared.


Fabrizio86:
Screenshot 2026-09-13 at 7 39 01 AM
Screenshot 2026-09-13 at 7 39 15 AM

michaelsakharov:
Hi @Fabrizio86 thanks for putting time into this, However, I'm going to close it without merging.

Mainly because it duplicates work already in progress and pretty much done #335
All the Review changes requested in #335 have already been addressed so this PR Addresses nothing not already addressed, and im already in the process of merging #335 as I write this.
This PR also reimplements the same components under the same names (`NavMeshSurface`, `NavMeshAgent`, `NavMeshObstacle`, `NavMeshLink`, `NavMeshModifier`, `NavMeshModifierVolume`, `NavMeshAreas`, `NavMeshAgentTypes`) without referencing that PR, so the two can't coexist. 
If you're interested in navigation, reviewing or contributing to #335 would be really welcome.

There is also a few issues with this PR at first glance.

1. The runtime test suite goes from about 13 seconds to about 4 minutes
2. The Flow Field isn't a real flow field. It stores one direction per polygon, pointing from its centre to its parent polygon's centre, which can point into walls. It has no field.
3. Area costs are off by one, every area is charged the cost of the area below it
4. RebuildTiles writes into a shared asset instance
5. every idle fixed tick is treated as a change

The commit message also mentions decompiling Detour's `FindPath`.
The Recast/Detour source we use is available in Anthology.

Thanks again for your interest in Prowl. 
Smaller, focused PRs that build on the existing work are much easier for us to review and merge.

